PR #149, opened 2019-10-18
improved return type generation

Kharhamel:


generated/gmp.php
```diff
@@ -78,11 +78,12 @@ function gmp_import(string $data, int $word_size = 1, int $options = GMP_MSW_FIR
  * @throws GmpException
  *
  */
-function gmp_random_seed($seed): void
+function gmp_random_seed($seed): \GMP
```
Kharhamel: incorrect?

generated/image.php
```diff
@@ -434,13 +434,14 @@ function imagecolormatch($image1, $image2): void
  * @throws ImageException
  *
  */
-function imageconvolution($image, array $matrix, float $div, float $offset): void
+function imageconvolution($image, array $matrix, float $div, float $offset)
```
Kharhamel: incorrect?

PR #154, opened 2019-11-04
Update rector-migrate.yml to support Rector 0.5

radimvaculik:
https://github.com/rectorphp/rector/blob/master/docs/AllRectorsOverview.md#renamefunctionrector


Kharhamel:
Thanks you.

PR #156, opened 2019-11-30
Fix PHP 7.4 warning for invalid array access

andreasschroth:
This warning is easily reproducable under PHP 7.4 with code like this:

```php
\Safe\strtotime('nonsense');
```

error_get_last() might return null and then an invalid array access occurs.
For more information see this RFC: https://wiki.php.net/rfc/notice-for-non-valid-array-container

### DIFF
--- a/lib/Exceptions/AbstractSafeException.php
+++ b/lib/Exceptions/AbstractSafeException.php
@@ -8,6 +8,10 @@ abstract class AbstractSafeException extends \ErrorException implements SafeExce
     public static function createFromPhpError(): self
     {
         $error = error_get_last();
+        
+        if ($error === null) {
+            return new static();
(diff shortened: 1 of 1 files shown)


Kharhamel:
Is strtotime the only function that can create this bug? If yes maybe we should edit this function rather than the exception system?

andreasschroth:
I'm not absolutely sure tbh, but even if that's the case I think it should be fixed that way - if `$error` is null, it doesn't make sense to try to access it like an array.

Kharhamel:
But then why do we throw an exception if $error is null? Maybe it should throw a different kind of exception, or even nothing at all.

Kharhamel:
the fix was implemented in #170

PR #158, opened 2019-12-05
Support for php 7.4

snapshotpl:


szepeviktor:
@snapshotpl Already included
https://github.com/thecodingmachine/safe/blob/master/.travis.yml#L10-L12

Kharhamel:
Yes. We put the 7.4 version in the 'allowed_failure' because it looks like the 7.4 travis pipeline isn't stable yet.

PR #161, opened 2019-12-12
Password hash

Kharhamel:


generated/password.php
```diff
@@ -97,7 +97,7 @@
  * algorithm, will result
  * in the password parameter being truncated to a
  * maximum length of 72 characters.
- * @param int $algo A password algorithm constant denoting the algorithm to use when hashing the password.
+ * @param int|string $algo A password algorithm constant denoting the algorithm to use when hashing the password.
```
bdsl: The `\PASSWORD_DEFAULT` constant is equal to null in PHP 7.4, so:

```suggestion
 * @param int|string|null $algo A password algorithm constant denoting the algorithm to use when hashing the password.
```

But since this is in the 'generated' directory, doesn't the change need to be made upstream in the PHP docs xml files, and then imported into here?

Kharhamel: we can use generator/src/PhpStanFunctions/CustomPhpStanFunctionMap.php to override the phpstan docs which we use to generate the types

bdsl: @Kharhamel what do you think about adding the `|null`?

Kharhamel: done

bdsl: Thanks. This should resolve issue that @davidsneighbour found.

PR #166, opened 2019-12-16
Fix generator phpstan regexp

szepeviktor:
a lot of _ignored by mistake_ errors popped out
https://travis-ci.org/thecodingmachine/safe/jobs/625615584#L298

### DIFF
--- a/generator/phpstan.neon
+++ b/generator/phpstan.neon
@@ -1,6 +1,6 @@
 parameters:
     ignoreErrors:
-        - "#subject of function str_replace expects array|string, string|true given#"
-        - "#createExceptionFile\\(\\) expects string, int|string given#"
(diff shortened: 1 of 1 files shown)


Kharhamel:
I am not sure I understand. Do you have theses errors after editing the phpstan.neon or before?

szepeviktor:
After *fixing* the broken regexp-s.
It covered all string related type errors: the `|` sign was not escapes.

szepeviktor:
Now we face all errors introduced while using broken regexp-s.

Kharhamel:
ok I get it, you are right the regexp are false and cause the phpstan to ignore too many errors. I am going to fix them.

Kharhamel:
Done: #168

Kharhamel:
This is merged. Thank you for pointing this out.

PR #170, opened 2019-12-18
fixed incorrect excecption creation

Kharhamel:


snapshotpl:
@Kharhamel I think that this fix really need test case

PR #173, opened 2019-12-20
Delete references to pending issues before releasing 1.0 from readme

bdsl:
Yes, technically 0 is a number, but we don't need to mention it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,10 +162,6 @@ Safe is loading 1000+ functions from ~85 files on each request. Yet, the perform
 In case you worry, using Safe will "cost" you ~700µs on each request. The [performance section](performance/README.md) 
 contains more information regarding the way we tested the performance impact of Safe. 
 
-## Work in progress
-
-There are a number of pending issues [before releasing 1.0](https://github.com/thecodingmachine/safe/milestone/1)
-
 ## Learn more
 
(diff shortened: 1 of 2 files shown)


bdsl:
I'm not sure why the checks have failed on this change.

Kharhamel:
The ci always tries to regenerate the files and then compare to what is commited. The php doc has been updated recently which cause a change in the doc block of the function. 

You need to update the doc of you local project before regenerating the files (check CONTRIBUTING.md to see how to do all of this).
I can do it myself if it is not practical for you

bdsl:
Ok, I've run the generator and commited the change.

I wonder if there's interest in making this proccess more automated - I might be able to work on that. Potentially you could have a bot to submit PRs when the upstream docs change, or remove the generateds stuff from source control entirely and make it just part of a build. Maybe have two github repops, one manually edited without the generated stuff, and one auto generated based on a build from the first one and including the generated code for people to download and use.

Kharhamel:
I don't know. I like that it is still commited and has to be manually regenerated by a person, this makes it easier to track who modified what. You also need to have a local copy of the generated files to run your tests locally.

Kharhamel:
Also, people than require the library via composer don't download the generator code since it is labeled as export-ignore in .gitattributes

PR #177, opened 2020-01-03
Enhancement: Sort function names

localheinz:
This PR

* [x] sorts function names
* [ ] regenerates files

PR #178, opened 2020-01-03
Fix: Regenerate files

localheinz:
This PR

* [x] regenerates files

Also see #176 and #177.


localheinz:
Thank you, @Kharhamel!

Issue #179, opened 2020-01-06
\Safe\DateTimeImmutable comparison not correct


kunicmarko20:
```php
<?php

require_once __DIR__.'/vendor/autoload.php';

$timeLimit = \DateInterval::createFromDateString('2 hours');

$safeDateTime = new \Safe\DateTimeImmutable();
$phpDateTime = new \DateTimeImmutable();

\var_dump($safeDateTime->modify('+3 hours') < $safeDateTime->add($timeLimit)); // true
\var_dump($phpDateTime->modify('+3 hours') < $phpDateTime->add($timeLimit)); // false
```

Given this snippet, I would expect that both are `false`, but when using `\Safe\DateTimeImmutable` I am getting `true`.

kunicmarko20:
Fixed #180

PR #180, opened 2020-01-06
fixed an issue with datetimeimmutable comparaison

Kharhamel:
This pr solves #179

### DIFF
--- a/lib/DateTimeImmutable.php
+++ b/lib/DateTimeImmutable.php
@@ -28,14 +28,14 @@ class DateTimeImmutable extends \DateTimeImmutable
      */
     public function __construct($time = 'now', $timezone = null)
     {
-        parent::__construct();
+        parent::__construct($time, $timezone);
         $this->innerDateTime = new parent($time, $timezone);
     }
 
     //switch from regular datetime to safe version
     private static function createFromRegular(\DateTimeImmutable $datetime): self
     {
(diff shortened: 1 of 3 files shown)


lib/DateTimeImmutable.php
```diff
@@ -28,14 +28,14 @@ class DateTimeImmutable extends \DateTimeImmutable
...
-        parent::__construct();
+        parent::__construct($time, $timezone);
         $this->innerDateTime = new parent($time, $timezone);
     }
 
     //switch from regular datetime to safe version
     private static function createFromRegular(\DateTimeImmutable $datetime): self
     {
-        $safeDatetime = new self();
+        $safeDatetime = new self($datetime->format('Y-m-d H:i:s'), $datetime->getTimezone()); //we need to also update the wrapper to not break the operators '<' and '>'
```
bendavies: what about microseconds?

kunicmarko20:
@Kharhamel thank you for fast response, can we potentially get a patch release with this?

lib/DateTimeImmutable.php
Kharhamel: damn i forgot about theses

kunicmarko20: So milliseconds can't be used with < 7.4 because `::createFromFormat` doesn't accept it, microseconds should be used.

PR #183, opened 2020-01-08
Include microseconds in Date objects

### LINKED ISSUE #182: datetimeimmutable internal value doesn't use milliseconds
oversight in #180. $datetime->format('Y-m-d H:i:s') needs to also include milliseconds

kunicmarko20:
closes #182


kunicmarko20:
So milliseconds can't be used with < 7.4 because `::createFromFormat` doesn't accept it, will switch to microseconds.

Kharhamel:
Thanks a lot.

kunicmarko20:
Thank you for quick release!

PR #184, opened 2020-01-10
removed a useless composer dependency

Kharhamel:


localheinz:
@Kharhamel 

Perhaps you might like https://github.com/composer-unused/composer-unused?

Kharhamel:
On a big project with a lot of dependency, sure. On this one I am not sure.
It doesn't really apply anyway since this dependency was actually used by some dead code. I don't think composer-unused would have catch that.

PR #195, opened 2020-02-14
improved the continuous-integration workflow

Kharhamel:


Kharhamel:
@localheinz  downloading the php doc with svn takes a lot of time (any idea why?) but is needed in order to use both ./safe.php generate and phpunit.

Maybe cache the whole php doc then configure another action to sometimes update this cache?

localheinz:
@Kharhamel 

Perhaps #196 helps?

PR #196, opened 2020-02-14
Fix: Update PHP documentation on cache hit

localheinz:
This PR

* [x] updates the PHP documentation with `svn` on a cache hit

Follows #195.

💁‍♂ The initial checkout took 6 minutes - but I assume that updates will run faster, now that we have a primed cache.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -7,14 +7,14 @@ on:
   - "push"
 
 jobs:
-  
+
   init-php-doc:
     name: "Init the php documentation cache"
     runs-on: "ubuntu-latest"
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v2.0.0"
-      
+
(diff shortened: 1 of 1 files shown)


Kharhamel:
Weird, the cache should already be primed from my previous PR. How long does the cache live? If it is a short ttl, then we don't really need to do a svn update.

Kharhamel:
About your suggestion to use a github mirror for the doc, I am not sure. It probably isn't updated as much as the svn repo.

What I would like to do would be to entirely take the svn download out the CI workflow. Maybe have another workflow triggered by a cron everyday at 1 am which will do the svn update and prime the cache. 

But that sounds difficult because we cannot share an action cache between 2 different workflows.

Kharhamel:
I spoke with @moufmouf  and we are going to use the github mirror instead

localheinz:
Sounds good, @Kharhamel!

PR #201, opened 2020-02-25
Add execute permissions on update.sh

simPod:


Kharhamel:
Good catch, didn't think about the permissions, thank you

PR #207, opened 2020-03-13
Enhancement: Use actions/checkout to check out documentation

localheinz:
This PR

* [x] uses `actions/checkout` to check out the documentation


Kharhamel:
What does this change compared to the previous version?

localheinz:
@Kharhamel 

It's faster, and it does not depend on `update.sh`.

localheinz:
Build failure is fixed in #211.

localheinz:
Thank you, @Kharhamel and @OskarStark!

PR #204, opened 2020-03-06
Update generated files

simPod:

PR #208, opened 2020-03-13
Fix: Regenerate files

localheinz:
This PR

* [x] regenerates files

Potentially replaces #204, which has become outdated.

❗️ Blocked by https://github.com/thecodingmachine/safe/pull/209.


localheinz:
We have a 🐔 and 🥚 problem, so I'm regenerating files in #209 instead.

PR #226, opened 2020-06-29
fix #225

### LINKED ISSUE #225: Errors when running on PHP8
When running the project (throuh the infection package) on PHP8, I get this error:

```
PHP Warning:  "resource" is not a supported builtin type and will be interp  
reted as a class name. Write "\Safe\resource" or import the class with "use  
" to suppress this warning
```

It looks like the `resource` typehint in the `generated/sockets.php` file in the
`socket_wsaprotocol_info_export` method is not supported in PHP8 and should
thus be removed to make this library and infection compatible with PHP8.

fezfez:


WouterSioen:
Yes please!

Kharhamel:
Done, sorry i didn't the project in a while. I will do a release shortly

PR #259, opened 2020-12-24
Enclose function with an existence test

### LINKED ISSUE #253: Cannot redeclare Safe\array_combine()
Hi there, congrats on the project, I just discovered it and I think is great.

Now, I've tried to add it to one of my libraries and now when I run phpstan (with our without the rule) I get the following error:
```
PHP Fatal error:  Cannot redeclare Safe\array_combine() (previously declared in /home/user/project/vendor/thecodingmachine/safe/generated/array.php:20) in /home/user/project/vendor/thecodingmachine/safe/generated/array.php on line 20

Fatal error: Cannot redeclare Safe\array_combine() (previously declared in /home/user/project/vendor/thecodingmachine/safe/generated/array.php:20) in /home/user/project/vendor/thecodingmachine/safe/generated/array.php on line 20

In array.php line 20:
                                                                                                                                                              
  Cannot redeclare Safe\array_combine() (previously declared in /home/user/project/vendor/thecodingmachine/safe/generated/array.php:20)  
                                                                                                                                                              
```

The file obviously don't have the function declared twice and I'm completely lost as for where it could be coming from. Any ideas? I'm on 1.3.3.

Thanks!

asbiin:
This PR creates a enclose test for every function, to test if the function already exists or not.
Example
```php
if (! function_exists('curl_copy_handle')) {
    /**
     * Copies a cURL handle keeping the same preferences.
     * 
     * @param \CurlHandle $handle A cURL handle returned by
     * curl_init.
     * @return resource Returns a new cURL handle.
     * @throws CurlException
     * 
     */
    function curl_copy_handle(\CurlHandle $handle)
    {
        error_clear_last();
        $result = \curl_copy_handle($handle);
        if ($result === false) {
            throw CurlException::createFromPhpError();
        }
        return $result;
    }
}
```

This fix #253
It will also fix https://github.com/nunomaduro/larastan/issues/741


Kharhamel:
Sorry if I didn't see your PR sooner, I was away during the end of year.

I think this kind of fix is completly overkill compared to the specific issue of array_combine. So many 'if' and 'function_exist' that will have to be runned for every request has to be a major performance hit, which would be pretty embarrasing for a utility library like Safe.

Again sorry for closing your PR, but there has to be a better way. I will have to make a lot of changes anyway to this project to accomodate for php8.

asbiin:
OK I understand what you mean.
[Laravel helper](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Support/helpers.php) is doing it like that which was the reason I've done it this way.

PR #260, opened 2020-12-28
Fix generator

asbiin:
This fix generate the functions with the last data.
Recent changes in https://github.com/php/doc-en have been made, especially on the variadic arguments.

You will find multiple changes here:
- trim phpdoc: this remove trailing right space on the phpdoc
- add a `position` parameter to `PhpStanFunction::getParameter` so we can find the right parameter using the position, instead of the name, because sometimes parameter name differs between phpstan and php/doc-en.
- change the way to detect variadic parameters, since it has [change in php/doc-en](https://github.com/php/doc-en/commit/9e0f03ac354d797d1d16c0fcc1663e5e170f2727)
- fix com_load_typelib and sem_get function definition, because their definition is wrong in phpstan
- add a test in MethodTest


Kharhamel:
Thanks a lot! The only issue with this pr is that a lot of functions have been remove from safe, so if I merge this, I will have to make a new major version.
I plan on doing a v2.0 anyway, but I plan on making it only support php8 and above.
Are you ok if theses changes won't be available for php7?

Kharhamel:
@asbiin Just pinging you to make sure you are ok with what I said.

asbiin:
Thanks to reach out @Kharhamel !
I'm fine with you yes, I don't mind if these changes of not included in php7.

Anyway, what do you think about the fixes on the generator itself, are they relevant?

Kharhamel:
Not sure I get the purpose of the parameter $position you introduced but if no tests were broken I am fine with it.

asbiin:
I'll explain: the `Parameter` object is used to match the parameter between php/doc-en and phpstan declaration. The name of the parameter is used to match the parameters, but some times it does not work. The "position" is the position of the parameter in the declaration, to have a better match.

asbiin:
@soullivaneuh It's not mine to decide, but I agree with it ;) @Kharhamel maybe?

PR #261, opened 2020-12-28
Fix generator and update deps

asbiin:
This is the same as #260, with an addition of updating the dependencies in the generator.


Kharhamel:
duplicated with #260, I will update the dependencies in another PR

Issue #262, opened 2021-01-05
It looks like the function current is missing


bilelgasmi:
Hi, it looks like [current()](https://www.php.net/manual/en/function.current.php) is missing.

Any plans to add it? Thanks!

shish:
This family of functions can return `false` because they are empty, or can return `false` because the element contains an element with the value `false`, so one can't spot the error by looking just at the return value.

We could potentially add special-cases for this family of functions where we first check the length of the array, throw an exception if the array is empty, and then go on to return whatever the function returns, but that's slightly less-trivial

PR #263, opened 2021-01-05
[WIP] Add several functions

bilelgasmi:
- current (https://github.com/thecodingmachine/safe/issues/262)
- next
- prev
- end
- reset


Dgame:
This cannot work because it is possible and valid that e.g. `current` returns `false`:
```php
$arr = [false];
$value = current($arr);
```

shish:
Closing out old PRs - if this is still a desired function, it should probably be done by editing either the generator or special-cases file, rather than by editing the generated files

PR #269, opened 2021-03-07
Do not overwrite results in mb_parse_str

szepeviktor:
Use a different variable for success.
https://www.php.net/manual/en/function.mb-parse-str.php

### DIFF
--- a/generated/mbstring.php
+++ b/generated/mbstring.php
@@ -456,8 +456,8 @@ function mb_ord(string $string, string $encoding = null): int
 function mb_parse_str(string $string, ?array &$result): void
 {
     error_clear_last();
-    $result = \mb_parse_str($string, $result);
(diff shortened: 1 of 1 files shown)


dbrekelmans:
Hi @szepeviktor. While I understand the issue you're trying to fix, this is not the right solution. The file you changed is generated based on the phpdocs. You need to find and resolve the issue in the generator itself. Let me know if you need some pointers!

szepeviktor:
Thank you!

https://github.com/thecodingmachine/safe/blob/e454a3142d2197694d1fda291a4462ccd3f66e12/generator/src/WritePhpFunction.php#L79

Would `$safeResult` be okay with you?

szepeviktor:
Done in the next PR.

PR #273, opened 2021-03-20
fix DateTimeImmutable php 8 compatibility

### LINKED ISSUE #250: Safe\DateTimeImmutable is incompatible with PHP 8
```php
\Safe\DateTimeImmutable::createFromFormat(
    DateTime::ATOM,
    '2020-04-26T07:32:25+00:00'
);
```
Results in the following on PHP 8:
```
Error: Call to a member function format() on null
vendor/thecodingmachine/safe/lib/DateTimeImmutable.php:75
vendor/thecodingmachine/safe/lib/DateTimeImmutable.php:38
vendor/thecodingmachine/safe/lib/DateTimeImmutable.php:64
```
Tested on v1.3.2.

janatjak:
fix #250

### DIFF
--- a/lib/DateTimeImmutable.php
+++ b/lib/DateTimeImmutable.php
@@ -35,6 +35,12 @@ class DateTimeImmutable extends \DateTimeImmutable
     //switch between regular datetime and safe version
     public static function createFromRegular(\DateTimeImmutable $datetime): self
     {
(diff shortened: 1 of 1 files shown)


dbrekelmans:
Thanks @janatjak, this has been fixed by #278

PR #276, opened 2021-04-15
Change PHP requirement to prevent errors in PHP 8

tjveldhuizen:
Because the current code results in errors (at least when using Safe\DateTimeImmutable) in PHP 8, this package should not advertize compatibility with PHP 8.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -100,7 +100,7 @@
         ]
     },
     "require": {
-        "php": ">=7.2"
+        "php": "^7.2"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


jdreesen:
I'm not sure if this will help, as composer would probably install `v1.3.3` on PHP 8 because that's already released with `"php": ">=7.2"` requirement.

dbrekelmans:
Thanks @tjveldhuizen, I think @jdreesen is right. We're working on a PHP8 compatible release that will hopefully solve your issues :)

PR #278, opened 2021-06-15
Fix createFromFormat for PHP 8

Jean85:
This should be a transparent fix for #250; I understand that this library should be abandoned, but this fix will allow us to use it under PHP 8.0 without this bug, buying us time to build a new major (or Safe8).

### DIFF
--- a/lib/DateTime.php
+++ b/lib/DateTime.php
@@ -24,7 +24,7 @@ class DateTime extends \DateTime
      */
     public static function createFromFormat($format, $time, $timezone = null): self
     {
-        $datetime = parent::createFromFormat($format, $time, $timezone);
+        $datetime = \DateTime::createFromFormat($format, $time, $timezone);
         if ($datetime === false) {
             throw DatetimeException::createFromPhpError();
         }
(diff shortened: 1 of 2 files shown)


snapshotpl (approved):

KevinVanSonsbeek (approved):

dbrekelmans:
Thanks @Jean85

bendavies:
@Jean85 
> I understand that this library should be abandoned

can you explain this, please? which library?

Jean85:
@bendavies I was referring to the warning atop the readme, which was removed here: https://github.com/thecodingmachine/safe/commit/56ab9335d67c505d6eeae411584e2066e7a1ac0e#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L12-L26

https://github.com/thecodingmachine/safe8 was created for it, but I don't know what will happen... maybe they chose to not deprecate this lib?

dbrekelmans:
Minds were changed indeed and we're now working on a new release for the current library to support PHP8 :) @Jean85 @bendavies

PR #284, opened 2021-07-28
Fix PHP8 build

dbrekelmans:


dbrekelmans:
Still have to figure out why the generator is doing some weird things, like removing the `spl_autoload_register` method

Kharhamel (approved):

PR #288, opened 2021-08-15
Stop excluding functions with callable parameter synopsis

Duroth:
The generator checks for overloaded methods (functions with multiple signatures) and excludes these from the auto-generated functions.

However, this check errors on functions that include a callable in their parameters; in many cases, the callable's signature gets a synopsis of its own on the docs, which means the generator erroneously interprets it as an overloaded function.

This PR aims to fix this edge case by only pulling the synopsis from the "Description" section of the docpage, and excluding the "Parameters" part. This should mean that any synopses pertaining to the function's parameters, rather than the function itself, are now skipped and no longer trip up the generator.


generated/image.php
```diff
@@ -1772,33 +1772,13 @@ function imagegrabscreen()
...
-/**
- * Grabs a window or its client area using a windows handle (HWND property in COM instance)
- *
- * @param int $handle The HWND window ID.
- * @param bool $client_area Include the client area of the application window.
- * @return resource Returns an image object on success, FALSE on failure.
- * @throws ImageException
- *
- */
-function imagegrabwindow(int $handle, bool $client_area = false)
```
dbrekelmans, line 1784: This should not have been removed: https://www.php.net/manual/en/function.imagegrabwindow

generated/simplexml.php
```diff
@@ -1,87 +0,0 @@
...
- * @param \DOMNode $node A DOM Element node
- * @param string $class_name You may use this optional parameter so that
- * simplexml_import_dom will return an object of
- * the specified class. That class should extend the
- * SimpleXMLElement class.
- * @return \SimpleXMLElement|false Returns a SimpleXMLElement.
- * @throws SimplexmlException
- *
- */
-function simplexml_import_dom(\DOMNode $node, string $class_name = SimpleXMLElement::class): \SimpleXMLElement
```
dbrekelmans, line 21: This file should not have been removed. If the root cause is not related to the issue you're fixing, you could manually add it back (the build will fail, but I can still merge it and look at it in a different PR)

generated/spl.php
```diff
@@ -4,69 +4,6 @@
...
- * This function returns an array with the names of the interfaces that the
- * given class and its parents implement.
- *
- * @param object|string $class An object (class instance) or a string (class or interface name).
- * @param bool $autoload Whether to call __autoload by default.
- * @return array An array on success.
- * @throws SplException
- *
- */
-function class_implements($class, bool $autoload = true): array
```
dbrekelmans, line 17: This file should not have been removed

generated/url.php
```diff
@@ -28,35 +28,6 @@ function base64_decode(string $string, bool $strict = false): string
...
- * @param bool $associative If the optional associative parameter is set to true,
- * get_headers parses the response and sets the
- * array's keys.
- * @param resource $context A valid context resource created with
- * stream_context_create.
- * @return array Returns an indexed or associative array with the headers.
- * @throws UrlException
- *
- */
-function get_headers(string $url, bool $associative = false, $context = null): array
```
dbrekelmans, line 45: This file should not have been removed

generator/src/DocPage.php
```diff
@@ -163,9 +163,17 @@ public function getMethodSynopsis(): array
         if ($file === false) {
             throw new \RuntimeException('An error occured while reading '.$this->path);
         }
-        if (!preg_match_all('/<\/?methodsynopsis[\s\S]*?>[\s\S]*?<\/methodsynopsis>/m', $file, $functions, PREG_SET_ORDER, 0)) {
+
+        // Only evaluate the synopsis inside the `<refsect1 role="description">...</refsect1>` section of the doc page.
+        // Other synopses might occur in the `<refsect1 role="parameters">...</refsect1>` section, but these describe
+        // handlers, callbacks, and other callable-type arguments, not the function itself.
```
dbrekelmans, line 169: Nice fix! This adds a few functions that were not generated before :+1:

rector-migrate.php
```diff
@@ -372,7 +369,6 @@
             'imagegd2' => 'Safe\imagegd2',
             'imagegif' => 'Safe\imagegif',
             'imagegrabscreen' => 'Safe\imagegrabscreen',
-            'imagegrabwindow' => 'Safe\imagegrabwindow',
```
dbrekelmans, line 375: Keep in mind, if you revert unjust removals to also revert those function removals in this file

dbrekelmans (changes requested):
Thanks @Duroth! There are some removals by the generator which might or might not be related to your change. If they are not related, you can revert the removals. The build will fail, but I can still merge it and look for the root cause through a different PR.

generated/dom.php
```diff
@@ -11,7 +11,7 @@
  * as a native DOMElement node.
  *
  * @param \SimpleXMLElement $node The SimpleXMLElement node.
- * @return \DOMElement The DOMElement node added.
+ * @return \DOMElement|false The DOMElement node added.
```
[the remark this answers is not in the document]
Duroth: You can blame phpstan for this. ;)

generated/array.php
```diff
@@ -47,7 +47,7 @@ function array_combine(array $keys, array $values): array
  * @throws ArrayException
  *
  */
-function array_flip(array $array): array
+function array_flip($array): array
```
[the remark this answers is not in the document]
Duroth: Probably due to phpstan changing the parameters; PHP-Safe borrows parameter info from phpstan, and array_flip is one of the only functions I can find that is mapped with typed arrays:

```
'array_flip' => ['array', 'input'=>'array<int|string>'],
```

Just about all other functions use a generic `array` type hint.

Duroth:
- I've reverted the `.gitignore` changes;
- I've also reverted the generated code. While this may go against contributing guidelines, the (already) failing build is a good reminder of what functions still need to be looked at. I feel there's plenty more to be done before we can confidently re-generate a new baseline source.

dbrekelmans (approved):

dbrekelmans:
Thansk @Duroth!

PR #292, opened 2021-09-03
Added submodules, fixes #291

### LINKED ISSUE #291: Which version of php/doc-en was used?
I am trying to regenerate the latest version `1.3.3` and I have noticed a big problem: I don't know which version of [php/doc-en](https://github.com/php/doc-en) was used.
The problem with [php/doc-en](https://github.com/php/doc-en) is that no tags are used, it is constantly evolving and correctly documents the current version of PHP but not the previous ones (for example I wanted to try to fix the #289 problem but I cannot regenerate version `1.3.3`).
I therefore recommend using git submodules to keep track of which specific commit is used for each version of Safe.

Is there any trace somewhere of which [php/doc-en](https://github.com/php/doc-en) commit was used for each specific tag (at least the latest `1.3.3`) of Safe in order to reproduce it correctly?

fox91:


dbrekelmans:
Hi @fox91, thanks for the PR! I like the idea of submodules in this case.

If you resolve the conflicts, I will let the CI jobs run

fox91:
> Hi @fox91, thanks for the PR! I like the idea of submodules in this case.
> 
> If you resolve the conflicts, I will let the CI jobs run

This PR is supposed to be merged in a new branch created after tag `v1.3.3` (as I done in my fork).
This way is possible to solve other bug and maintain support to PHP 7.3+

After that you can also merge with `main` branch (overwrite `generator/doc/update.sh` with the new one).

dbrekelmans:
Hi @fox91, sorry for the late response, but I've merged this to a new `1.x` branch. Any future bugfixes regarding PHP7.x can be targeted against this branch and it will use the specific php-docs commit that you have introduced here.

Thanks for the contribution!

PR #304, opened 2021-10-17
Added PHPStan-generics to preg_replace

tjveldhuizen:
The subject in preg_replace can be a string or an arrays of strings. The return value is the same type as the provided subject. By adding the generics PHPStan 'knows' the correct return type in a situation like this:

```php
use function Safe\preg_replace;

$foo = 'some string 123';
$bar = preg_replace('/[0-9]/', '', $foo);
$baz = trim($bar); // PHPStan complains that trim does not accept string|string[]
```

### DIFF
--- a/lib/special_cases.php
+++ b/lib/special_cases.php
@@ -152,6 +152,9 @@ function apcu_fetch($key)
  *
  * @throws PcreException
  *
+ * @phpstan-template T of array|string
+ * @phpstan-param T $subject
+ * @phpstan-return T
  */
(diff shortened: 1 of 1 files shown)


ZebulanStanphill:
@tjveldhuizen [PHPStan 1.0](https://phpstan.org/blog/phpstan-1-0-released) just came out, and it includes support for `@template T of array`, so once this project updates its PHPStan requirement, this PR can be rebased and should start working as intended.

Kharhamel:
Sorry I completly missed this PR. I will update the phpstan recs this week

PR #316, opened 2022-01-05
Fix fputcsv for PHP 8.0

Jean85:
Fixes #310

This is probably a temporary workaround, the definitive fix should be what proposed in https://github.com/thecodingmachine/safe/issues/310#issuecomment-1005535320

PR #318, opened 2022-01-05
Fix fputcsv annotation

Jean85:
During #316 I restricted the type of the `$field` argument, and that triggered my static analysis tools while importing the fix.

As shown in https://3v4l.org/ahdmn, the original `fputcsv` function will try to cast everything to `string`, so every scalar is safe, array are risky (they become `Array`), objects explode, with the exception of stringable ones.

### DIFF
--- a/generated/filesystem.php
+++ b/generated/filesystem.php
@@ -896,40 +896,6 @@ function fopen(string $filename, string $mode, bool $use_include_path = false, $
 
 
 /**
- * fputcsv formats a line (passed as a
- * fields array) as CSV and writes it (terminated by a
- * newline) to the specified file stream.
- *
- * @param resource $stream The file pointer must be valid, and must point to
- * a file successfully opened by fopen or
- * fsockopen (and not yet closed by
- * fclose).
- * @param array $fields An array of strings.
- * @param string $separator The optional separator parameter sets the field
- * delimiter (one single-byte character only).
- * @param string $enclosure The optional enclosure parameter sets the field
- * enclosure (one single-byte character only).
- * @param string $escape The optional escape parameter sets the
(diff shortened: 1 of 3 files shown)


Kharhamel:
This looks good. The CI error seems to come from a cache error (I had the same on another PR) so I think we can safely merge it. How sure are you of your fix?

Jean85:
Umh, the PR seems kinda messed up, because the fix was already merged in #316...
Maybe you erased that with your forced push?

Kharhamel:
What you mean? Which force push? I only ever force-push to my own branches, never to master.

I also fixed the cache, which should solve your CI error. You just need to pull master to benefit from the change.
Also could you squash all of your commits into a single one? That would look cleaner

Jean85:
GitHub doesn't agree... That's what I see:
immagine

You force-pushed, erasing https://github.com/thecodingmachine/safe/commit/6b3abbedb2571f4cfde4bf25b449279fb2a6b288 which was @dbrekelmans merging my #316 into master.

That's why this PR is messy: it's #316 + https://github.com/thecodingmachine/safe/pull/318/files/9f63c60d5f1e746873629fd5447e6b80f7622770..246d93cd3a7d52d7beb33028918d64de45a24329

Kharhamel:
That's weird. If it is really messed up, the easy way to fix is to simply destroy the commits (but not the changes), stash the changes, start over from a clean master branch and commit. So something like this:

`
git reset _hashOfTheCommitBeforeYourChanges_
git stash
git checkout master
git pull master
git checkout -b _newBranch_
git commit
`

Jean85:
Done!

Kharhamel:
Damn i just created a new release. I'll do another one i guess

PR #322, opened 2022-01-18
bring in curl handles

mmoll:
see https://php.watch/versions/8.0/resource-CurlHandle

### DIFF
--- a/generated/curl.php
+++ b/generated/curl.php
@@ -7,13 +7,13 @@ use Safe\Exceptions\CurlException;
 /**
  * Copies a cURL handle keeping the same preferences.
  *
- * @param resource $handle A cURL handle returned by
+ * @param \CurlHandle $handle A cURL handle returned by
  * curl_init.
- * @return resource Returns a new cURL handle.
+ * @return \CurlHandle Returns a new cURL handle.
  * @throws CurlException
  *
  */
-function curl_copy_handle($handle)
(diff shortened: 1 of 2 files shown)


mmoll:
Not sure if this is correct, but I saw errors like 
```
[...]
  90     Parameter #1 $handle of function Safe\curl_setopt expects resource, CurlHandle given.   
  92     Parameter #1 $handle of function Safe\curl_exec expects resource, CurlHandle given.     
  93     Parameter #1 $handle of function Safe\curl_getinfo expects resource, CurlHandle given.
```
also with #315.

mmoll:
now what? 🤔

dbrekelmans:
Hi @mmoll, sorry I missed this PR.

Thanks for the effort so far.

The correct way to fix this, is to add these functions to `generator/src/PhpStanFunctions/CustomPhpStanFunctionMap.php` as was done in #315.
`generated/curl.php` is, as the directory suggest, generated and should not be edited manually.
This allows us to automatically update the generated files when the php docs are updated instead of manually updating all the changed functions on every php release.

If you could update the PR and make sure the build is green, I will gladly merge this!

mmoll:
Thanks @dbrekelmans, I hope it's coming back green now. 💚

mmoll:
should be fixed now

dbrekelmans:
Thanks @mmoll!

PR #334, opened 2022-03-18
Replace `error_get_last()` with a custom error handler

### LINKED ISSUE #332: Take into account global error handlers
Hi there,

I was reading [this thread](https://twitter.com/another_clue/status/1503830167252307973) on Twitter by [Christian Lück](https://twitter.com/another_clue) and I immediately thought on Safe. 

I just checked and just like on his example, this library [uses](https://github.com/thecodingmachine/safe/blob/v2.1.2/generated/Exceptions/FilesystemException.php#L8) `error_get_last()` without accounting for any other global error handlers set by the application. I checked `fopen()` but I guess the implementation is the same for all functions using `error_get_last()`.

What do you think? Shall this be addressed? Would you like a MR?

j3j5:
Try to fix #332 

This is my first serious attempt at a PR on here, so please, any suggestion is welcome.

While doing this I realized the implementation for the exception handling on the curl module was wrong. With PHP8 the `$ch` parameter doesn't exist anymore and it's now `$handle`. I fixed that and added the case for the other two new handles `$multi_handle` and `$share_handle`.

PHPStan throws no errors on my machine but I didn't have time to test it properly, so more eyes are welcome before merging.


Kharhamel:
I am sorry, I didn't have the time to review your PR, i will check it this week.

I also pinged @moufmouf who is much more expert in error_handlers

j3j5:
I've rebased the branch and fixed the test. I'm updating the original description to mention the new PCRE Exception.

shish:
Reviving this project after a couple of years of inactivity, this PR looks like it _might_ still be useful so I'm going to leave it open, but marking it as a draft until it is rebased on current master

j3j5:
Given how long it has passed since I open this PR, I probably should just remake it instead of a simple rebase, I don't know why it's so big. I'll try to give it a try as soon as I have some free time.

OskarStark:
Feel free to recreate it if a rebase is too complex due to the project progress

mstrelan:
This would indeed still be useful.

Issue #348, opened 2022-04-25
New, autoloader-centric file structure


szepeviktor:
Aka. _for machines only_

```
repo/
├─ generated/
│  ├─ src/
│  ├─ functions.php
```

- all classes go (generated or copied) in `src` directory as per PSR-4
- all functions go (generated or copied) into `functions.php` file
- keep `generated/functionsList.php`
- deprecated things get a PHPDoc tag

```php
/**
 * @deprecated Reason, date, version ...
 */
```

Resulting Composer config.
```json
    "autoload": {
        "psr-4": {
            "Safe\\": [
                "generated/src/"
            ]
        },
        "files": [
            "generated/functions.php"
        ]
    },
```
All other directories could be excluded from distribution.

Would you support a PR that modifies the generator?

szepeviktor:
@Kharhamel I understand it may hurt your feelings.

szepeviktor:
Asked how to do it: https://github.com/rectorphp/rector/issues/7131

Kharhamel:
I am not sure I understand the link with rector but yes, that new structure looks good, i would definitly like a PR to do that, especially if it solves issues like #253.

However I am thinking really hard about generating several directories of functions, to be loaded depending of the php version of the client. Do you think we could do something like this:

```
repo/
├─ generated-lastest/
│  ├─ src/
│  ├─ functions.php
├─ generated-7.4/
│  ├─ src/
│  ├─ functions.php
```

And then the autoload config in composer would reference a single entrypoint file that would redirect us to the correct folder depending on the php version?

szepeviktor:
> to be loaded depending of the php version of the client

That is done by keeping one git branch for each PHP version.
At least many other packages do it this way.

szepeviktor:
> I am not sure I understand the link with rector

That Rector issue is about automatically combining many .php file into one `functions.php`.

shish:
As far as I'm aware this won't work with multiple PHP versions, because for example PHP8.4 includes functions which are syntactically invalid in 8.1 (with `true` as a return type). If I'm wrong and it can be done, I would welcome a proof-of-concept to see that it does actually work :)


> That is done by keeping one git branch for each PHP version.
> At least many other packages do it this way.

I've been hoping that this was possible, and searching for examples of people doing it, but been unable to find any approach which ticks all of the boxes we need :( Again, if anybody has specific examples and a proof-of-concept that'd be excellent <3

[...]

PR #350, opened 2022-04-28
Replace PSR-4 with classmap autoloader

### LINKED ISSUE #253: Cannot redeclare Safe\array_combine()

j3j5:
I am once again ~asking for your support~ trying to fix #253. 

I think this is the best approach (as discussed on [the phpdocparser issue](https://github.com/phpstan/phpdoc-parser/issues/117) ) and it's also much less disruptive than the renaming of all files done in #338.

I've tested the code with [this repo](https://github.com/j3j5/phpstan-larastan-bug), but please, test it yourselves to make sure I don't break anything else again :stuck_out_tongue: 

Hope this is the good one!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,13 +3,13 @@
     "description": "PHP core functions that throw exceptions instead of returning FALSE on error",
     "license": "MIT",
     "autoload": {
-        "psr-4": {
-            "Safe\\": [
-                "lib/",
-                "deprecated/",
-                "generated/"
-            ]
-        },
+        "classmap": [
+            "lib/DateTime.php",
+            "lib/DateTimeImmutable.php",
+            "lib/Exceptions/",
+            "deprecated/Exceptions/",
+            "generated/Exceptions/"
+        ],
         "files": [
             "deprecated/apc.php",
             "deprecated/array.php",


Kharhamel:
This looks really easy. I need to think about how to test this on a bigger scale. I will probably create a beta for this but this means i need to merge this into master.

j3j5:
Yeah, please double check that I didn't miss any class. The only downside is that if a new class is ever added that doesn't fall in the folders (new exceptions will be fine), it'll need to be added manually. I don't think there will be another case like Datetime in this library, but something to be aware of.

szepeviktor:
@j3j5 I think #348 is the ultimate solution.

j3j5:
As I mentioned on previous attempts, I don't mind which approach we use, I just want to get rid of the error so whatever is easier/better for the actual maintainers :)

Kharhamel:
@szepeviktor didn't propose his pull request yet, so I am going to merge this for now and create a beta to test it. Is it ok with you @j3j5?

j3j5:
Yeah, I mean, it should fix the problem for now (hopefully) and later we can go with the "final solution" if y'all like it better.

PR #357, opened 2022-05-19
Make Safe\Exceptions\JsonException inherits from \JsonException

jdecool:
Currently the `Safe\json_` functions throws a `Safe\Exceptions\JsonException` which is not related to the standard PHP `JsonException`.

So this PR add intheritance between `Safe\Exceptions\JsonException` and `\JsonException` to solve this issue.

### DIFF
--- a/lib/Exceptions/JsonException.php
+++ b/lib/Exceptions/JsonException.php
@@ -3,7 +3,7 @@
 
 namespace Safe\Exceptions;
 
-class JsonException extends \Exception implements SafeExceptionInterface
+class JsonException extends \JsonException implements SafeExceptionInterface
 {
     public static function createFromPhpError(): self
     {


Kharhamel:
Are you sure this doesn't break anything? I got some really bad suprises with some recent PRs

jdecool:
> Are you sure this doesn't break anything? I got some really bad suprises with some recent PRs

I don't think so.

Previously it throws an `\Exception`. Now it will throw an `\JsonException` which inherits from the standard `\Exception` (see https://www.php.net/manual/en/class.jsonexception).

So it shouldn't have any BCB.

Kharhamel:
That's what I though too, but then stuff like #253. I am still going to merge this because I don't think there is a way to anticipate this kind of issues anyway

PR #369, opened 2022-07-20
FIX: Deprecated array_replace and array_replace_recursive

Kharhamel:


shadowhand:
Is there some documentation that explains why some functions get deprecated?

JacekAndrzejewski:
> Is there some documentation that explains why some functions get deprecated? I'm all for native when possible, but I'd like to understand why Safe alternative functions are no longer recommended.

Looking at the code it seems that those could never throw, and looking at PHP docs it looks like those two functions always returned arrays. So this alternative was no really not safer in any way.

shadowhand:
@JacekAndrzejewski thanks!

Kharhamel:
> Is there some documentation that explains why some functions get deprecated? I'm all for native when possible, but I'd like to understand why Safe alternative functions are no longer recommended.

The whole library is synced with the official documentation, so when they randomly edit a function page to state that it doesn't return false/null anymore, we have to drop support for this function.
Bu since I try to respect semVer, i deprecate them instead.

PR #376, opened 2022-08-04
issue 324: change parameter name from options to flags

MarcinGladkowski:


dbrekelmans (approved):

danilopolani:
Hey @MarcinGladkowski and @dbrekelmans , this was a breaking change, but has been marked as a minor update. For example, it has broken the package https://github.com/jessarcher/laravel-castable-data-transfer-object [here](https://github.com/jessarcher/laravel-castable-data-transfer-object/blob/main/src/CastableDataTransferObject.php#L27).

Is there any chance we can rollback this?

dbrekelmans:
Hi @danilopolani. Thanks for the report! This is a though one, because it's hard to support BC for parameter names. This was one of the issues brought up in the internals discussion on the named arguments RFC. Some big open-source projects specifically state that they don't support named arguments as part of their BC promise.

While this library doesn't have a written policy on BC, I personally think it's a lot of effort to support this, probably more than it's worth. If you have strong opinions about this, please feel free to continue the discussion here!

danilopolani:
> Hi @danilopolani. Thanks for the report! This is a though one, because it's hard to support BC for parameter names. This was one of the issues brought up in the internals discussion on the named arguments RFC. Some big open-source projects specifically state that they don't support named arguments as part of their BC promise.
> 
> While this library doesn't have a written policy on BC, I personally think it's a lot of effort to support this, probably more than it's worth. If you have strong opinions about this, please feel free to continue the discussion here!

Definitely a tough one, thank you for the explanation 😄 Gonna submit a PR to that project to fix this behaviour, in the meanwhile I've solved by requiring a fixed (old) version of this package.

Thank you again!

lib/special_cases.php
```diff
@@ -27,15 +27,15 @@
...
  * @param int<1, max> $depth   User specified recursion depth.
- * @param int $options Bitmask of JSON decode options.
+ * @param int $flags Bitmask of JSON decode options.
  *
  * @return mixed
  * @throws JsonException if the JSON cannot be decoded.
  * @link http://www.php.net/manual/en/function.json-decode.php
  */
-function json_decode(string $json, bool $assoc = false, int $depth = 512, int $options = 0)
+function json_decode(string $json, bool $assoc = false, int $depth = 512, int $flags = 0): mixed
```
shadowhand, line 36: Just as a note, this could be considered a breaking change in PHP 8.x because anyone using the named argument will be faced with a breaking change in a minor package release.

generated/openssl.php
```diff
@@ -1159,34 +1159,6 @@ function openssl_public_encrypt(string $data, ?string &$encrypted_data, $public_
...
- * @param int $length The length of the desired string of bytes. Must be a positive integer. PHP will
- * try to cast this parameter to a non-null integer to use it.
- * @param bool|null $strong_result If passed into the function, this will hold a bool value that determines
- * if the algorithm used was "cryptographically strong", e.g., safe for usage with GPG,
- * passwords, etc. TRUE if it did, otherwise FALSE
- * @return string Returns the generated string of bytes on success.
- * @throws OpensslException
- *
- */
-function openssl_random_pseudo_bytes(int $length, ?bool &$strong_result = null): string
```
KevinVanSonsbeek, line 1179: I just ran into this breaking change when upgrading safe. Shouldn't this method have been moved to the deprecated space instead of just being removed? As i randomly ran into an error after doing a minor update, without a deprecation notice prior.

KevinVanSonsbeek: I'm up for making a PR is that is preferable btw

PR #395, opened 2022-09-29
stream_filter_append & stream_filter_prepend use mixed instead of array

ildyria:
https://www.php.net/manual/en/function.stream-filter-append.php
Defines `stream_filter_append` as :
```
stream_filter_append(
    resource $stream,
    string $filtername,
    int $read_write = ?,
    [mixed](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.mixed) $params = ?
): resource
```
This PR aims to fix the inconsistency as here `$params` is defined as `array`.

Also updates https://www.php.net/manual/en/function.stream-filter-prepend.php

### DIFF
--- a/generated/curl.php
+++ b/generated/curl.php
@@ -3109,17 +3109,6 @@ function curl_multi_setopt(\CurlMultiHandle $multi_handle, int $option, $value):
  *
  *
  *
- * CURLOPT_XFERINFOFUNCTION
- *
- * A callback accepting two parameters.
- * Has a similar purpose as CURLOPT_PROGRESSFUNCTION but is more modern
- * and the preferred option from cURL.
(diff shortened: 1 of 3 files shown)


Kharhamel:
You cannot directly update the generated files because the change will be lost the next time we will regenerate them. You need to update the generator instead.

To solve this, the simple solution is to edit this file: https://github.com/thecodingmachine/safe/blob/master/generator/config/CustomPhpStanFunctionMap.php
To make stream_filter_append return the type you want

ildyria:
@Kharhamel I updated as you advised.

Kharhamel:
Thank you for the MR. I will deploy this shortly

PR #405, opened 2022-12-26
Fix: the json_last_error will not reset when using JSON_THROW_ON_ERROR flag

m3m0r7:
The PHP has a problem that the json_last_error will not reset when turning on the JSON_THROW_ON_ERROR flag.

- see: https://github.com/php/php-src/issues/10166

When I call `\Safe\json_decode(...,JSON_THROW_ON_ERROR)` after `\json_decode("\00 invalid json")`, and then the `\Safe\json_decode()` function throws an exception always because json_last_error was not reset.

I fixed the problem with this pull request.
And I did add tests.


staabm:
I think this got fixed by https://github.com/thecodingmachine/safe/pull/451

PR #409, opened 2023-02-21
FIX: Fix openssl_pkey_get_* functions

martinssipenko:
- Fixed the input and output of the `openssl_pkey_get_private` function as starting from PHP 8.0 it accepts:
	- `\OpenSSLAsymmetricKey|\OpenSSLCertificate|array|string` for `$private_key`
	- `null|string` for `$passphrase`
and returns `OpenSSLAsymmetricKey|false`.

- Fixed the input and output of the `openssl_pkey_get_public` function as starting from PHP 8.0 it accepts:
	- `\OpenSSLAsymmetricKey|\OpenSSLCertificate|array|string` for `$public_key`
and returns `OpenSSLAsymmetricKey|false`.

- Added `openssl_pkey_get_details` function as it's actually falsy, but was not picked up by automatic code generation because of wording in PHP docs, but now is after [this change](https://github.com/php/doc-en/commit/0e80601cabf499c7db372149916b5cb3e2740f1b).

### DIFF
--- a/generated/openssl.php
+++ b/generated/openssl.php
@@ -1482,4 +1482,3 @@ function openssl_x509_read($certificate): \OpenSSLCertificate
     }
     return $safeResult;
 }
-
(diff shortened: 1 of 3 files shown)


Kharhamel:
Thanks a lot, I will try to deploy this this week end

PR #419, opened 2023-04-10
Use phpstan return logic to infer return type from argument

ildyria:
As migrating to php 8.2, I had to change `utf8_encode` (soon deprecated) to  `mb_convert_encoding`.
Unfortunately this function when given a string as input returns a string, and when given an array returns an array.
This fix aims to leverage on phpdoc `@return` logic to infer the correct return type.


shish:
Closing old PRs as lots has changed, feel free to re-open if this is still relevant

(FWIW I tested this in my current master branch, and it broke the generated code because it put `($string is string ? string : array)'` verbatim into the function's return type)

PR #420, opened 2023-04-12
change signatures for `openssl_sign` to Adapt to php8+

BadJacky:
In PHP8+, there is a new return type `OpenSSLAsymmetricKey` on the function `openssl_pkey_get_private` , but the `safe\openssl_sign` does not allow this type as the `private_key` parameter.

You can get details from the changed files and let me know if you have other questions.


BadJacky:
> You cannot juste edit the generated files, or your change will be overwritten when we generate from the doc. You need to edit the file CustomPhpStanFunctionMap.php to overwrite the typings

You are right! I have changed the CustomPhpStanFunctionMap.php and pushed it again.

BadJacky:
Hi @Kharhamel ,
I noticed that you're preparing for the V3.0 release and I have encountered some errors that seem to be similar to yours. After some investigation, I found that fixing the errors would require us to address the documentation, as the generator depends on it. The exceptions include, but are not limited to, issues with gmp.
I was wondering if you're currently working on fixing these errors? If not, I'd be happy to collaborate with you on resolving them.
I have attached the fixed results image and let me know if you encountered any questions and I'm glad to fix them with you!
Screenshot 2023-04-17 at 07 17 20
Screenshot 2023-04-17 at 07 19 24

shish:
This should be fixed at the generator level in #533

PR #428, opened 2023-10-23
Add safe inet_pton

### LINKED ISSUE #270: add inet_pton to have pair with inet_ntop
currently this lib has `inet_ntop` implementation, but there is no `inet_pton` implementation for this lib

simPod:
Resolves #270

(is not generated from docs)

### DIFF
--- a/lib/special_cases.php
+++ b/lib/special_cases.php
@@ -8,6 +8,8 @@
 namespace Safe;
 
 use Safe\Exceptions\FilesystemException;
+use Safe\Exceptions\NetworkException;
+use function error_clear_last;
 use const PREG_NO_ERROR;
 
 use Safe\Exceptions\MiscException;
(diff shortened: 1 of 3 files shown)


lib/special_cases.php
```diff
@@ -440,3 +442,18 @@ function fgetcsv($stream, int $length = null, string $separator = ",", string $e
     }
     return $safeResult;
 }
+
+/**
+ * @param string $ip A human readable IPv4 or IPv6 address.
+ * @return string Returns the in_addr representation of the given ip,
+ * @throws NetworkException
+ */
+function inet_pton(string $ip): string
```
staabm, line 451: do we know why we need a special case here?

staabm:
@simPod thanks for taking the time you put into it and making us aware of the problem

PR #431, opened 2023-10-23
Add custom base64_decode fcn mapping

simPod:
Phpstan defines multiple variants now which breaks it for Safe https://github.com/phpstan/phpstan-src/blob/80ce9a404787133ff3fa266f90602728bf1c349c/resources/functionMap.php#L410-L411


staabm:
thanks

PR #436, opened 2023-12-27
Fix `json_decode` `associative` argument consistency with PHP

dpi:
I like to use named arguments with `\json_decode()`, particularly with the `associative` argument.

e.g. `\json_decode($json, associative: true);`

Switching to use `thecodingmachine/safe` currently necessitates switching `associative` to `assoc`. This slows down migrations and introduces an inconsistency.

We should fix the named argument of `\Safe\json_decode` so it has the same name as PHP.


dpi:
Similar to #376

staabm:
cherry-picked with https://github.com/thecodingmachine/safe/pull/473

thank you

PR #456, opened 2024-11-21
PHP 8.4: Implicitly nullable parameters deprecated

valzargaming:
Implicitly marking parameters as nullable is deprecated, the explicit nullable type must be used instead.


valzargaming:
Should be good to test and merge now

staabm:
hey.

I am one of the new maintainers which recently got invited to revive this repo.

Thanks for the time you invested in this PR.
In the meantime we also merged other PRs with similar goals, so this PR is no longer required.

thanks again

PR #458, opened 2024-11-28
Add devcontainer config

shish:
(breaking shish/safe down into smaller parts for easier merging)

These configs make it so that one can open the Safe project in a compatible editor (eg vscode) and the IDE will do all the work (eg build, test) in a self-contained container with all the right versions of all the right tools installed. This is particularly useful for Safe, whose build process only runs with specific PHP versions, which may differ from what is installed on the developer's host system.


staabm:
thanks

PR #460, opened 2024-11-28
Update CI/CD workflows

shish:
(Splitting shish/safe into smaller parts for easier review)

This reworks the CI/CD workflow into smaller parts which run as a tree, so that a failure in one part doesn't block testing in a different unrelated part. Workflows will currently fail because of all sorts of brokenness in the code itself, but when the rest of the merge is complete, it should pass.

Also note that this tests PHP 8.1-8.4, even though shish/safe only tests 8.2-8.4 -- this is leaving room for somebody to implement php8.1 (which I couldn't figure out how to do), but if the other maintainers can't figure out php8.1 either, we can make an intentional decision to drop that later.


staabm:
thanks

PR #462, opened 2024-11-28
Use less-old phpstan

shish:
phpstan 1.5 is very out of date and the API for defining rules is incompatible with the current API.

tests are currently failing, but they're failing upstream and they should pass once all the other PRs are merged

PHPStan 2.0 was released a few days ago, but other parts of our dependency tree are locked to "1.X", so for now I'm just updating to "most recent 1.X" and we can look at 2.0 after the rest is stable.


staabm:
thanks

PR #463, opened 2024-11-28
Remove dev-master aliases

shish:
These are referring to very out-of-date branches and anybody who uses them is going to have a bad time


staabm:
thanks

PR #465, opened 2024-11-28
json_decode fixes

shish:
(Breaking the diff against shish/safe into smaller parts for easier reviewing)

### DIFF
--- a/lib/special_cases.php
+++ b/lib/special_cases.php
@@ -24,7 +24,7 @@ use Safe\Exceptions\SimplexmlException;
  * Wrapper for json_decode that throws when an error occurs.
  *
  * @param string $json    JSON data to parse
- * @param bool $assoc     When true, returned objects will be converted
+ * @param bool $associative     When true, returned objects will be converted
  *                        into associative arrays.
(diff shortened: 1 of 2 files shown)


lib/special_cases.php
```diff
@@ -33,10 +33,10 @@
  * @throws JsonException if the JSON cannot be decoded.
  * @link http://www.php.net/manual/en/function.json-decode.php
  */
-function json_decode(string $json, bool $assoc = false, int $depth = 512, int $flags = 0): mixed
+function json_decode(string $json, bool $associative = false, int $depth = 512, int $flags = 0): mixed
```
staabm, line 36: I think we should not rename the parameter, because of named arguments and php-src uses the name `$assoc`

staabm:
I will just merge https://github.com/thecodingmachine/safe/pull/451/files which is what I would expect.

thank you

staabm:
just found https://github.com/thecodingmachine/safe/pull/473/commits/faf662c1a8be86256e4449a1077aebb11a32b4e1, which tells me I was wrong as the param named changed in php-src.

will cherry pick

PR #466, opened 2024-11-28
Bring lib/ up to php8.4 standards

shish:
(Breaking the diff against shish/safe into smaller parts for easier reviewing)


staabm:
thanks

PR #476, opened 2024-11-28
Force generator to use PHP 8.1.x

staabm:
to improve DX, as before this PR the generator assumed composer would be run using a PHP 8.1 CLI.

lets define a platform version instead, so it no longer matters which PHP CLI version the developer is using to run composer


staabm:
//cc @shish

OskarStark (approved):

OskarStark:
Thanks

PR #481, opened 2024-12-02
enable strict_types in generator

shish:
(Continuing with pulling the changes from shish/safe into tcm/safe in small hopefully easy-to-review chunks)

### DIFF
--- a/generator/src/ComposerJsonEditor.php
+++ b/generator/src/ComposerJsonEditor.php
@@ -1,5 +1,6 @@
 <?php
 
+declare(strict_types=1);
 
 namespace Safe;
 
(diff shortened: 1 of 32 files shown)


staabm:
awesome, thanks for all this.

fyi, I would wait before merging new PRs, until [the generator is fixed](https://github.com/thecodingmachine/safe/pull/478#issuecomment-2511469302)

shish:
The generator is already working in my branch - these PRs are me trying to get us in-sync one small step at a time :)

staabm:
> The generator is already working in my branch - these PRs are me trying to get us in-sync one small step at a time :)

I see. do you think its possible to work on PRs which get the generator running first?

shish:
Technically anything is possible with enough effort - but those changes are built on top of these ones, and I worry that trying to re-order the commits would cause more merge-conflicts with more potential to mess things up ^^;;

staabm:
I see. please leave a hint in the PRs so we can be sure they get merged in the correct order.

PR #482, opened 2024-12-02
Constructor property promotions

shish:
(Pulling changes from shish/safe into smaller chunks for easier reviews)


staabm (approved):
thanks

PR #483, opened 2024-12-02
Fix regex matching for removed functions

shish:
(pulling fixes from shish/safe into tcm/safe one at a time)

### DIFF
--- a/generator/src/DocPage.php
+++ b/generator/src/DocPage.php
@@ -32,8 +32,8 @@ class DocPage
             }
         }
 
-        if (preg_match('/&warn\.removed\.function-(\d+-\d+-\d+)/', $file, $matches) && isset($matches[2])) {
-            $removedVersion = $matches[2];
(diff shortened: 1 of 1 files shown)


generator/src/DocPage.php
```diff
@@ -32,8 +32,8 @@ private function getIsDeprecated(string $file): bool
             }
         }
 
-        if (preg_match('/&warn\.removed\.function-(\d+-\d+-\d+)/', $file, $matches) && isset($matches[2])) {
-            $removedVersion = $matches[2];
+        if (preg_match('/&warn\.removed\.function-(\d+-\d+-\d+)/', $file, $matches)) {
```
staabm, line 35: this bug would have been found by PHPStan.

should we enable PHPStan coverage via `/phpstan.neon` for `generator/src/` ?

shish: Probably yes, but my top priority is getting shish/safe and tcm/safe in-sync so that we can get some green automated tests as a base to work from

staabm:
thanks

PR #484, opened 2024-12-02
Support more variations of doc phrasings

shish:
Eww at having to regex human-readable descriptions to figure this stuff out D:


staabm:
Is this something which can/should be covered by test-cases?

shish:
Actually I don't think this one is needed at all o.o

PR #486, opened 2024-12-02
Detect nullable-default parameters and generate appropriate outputs

shish:
This makes sure we are always generating explicitly-nullable parameters, even if the PHP documentation uses implicitly nullable (which is deprecated in php8.4)


staabm:
thanks

PR #487, opened 2024-12-02
Handle more PHPStan type edge-cases

shish:
Less crashing in code generation \o/


shish:
@staabm getting close, though it seems I prematurely assumed that all the PHP versions had the same test failures, when actually 8.1 and 8.2-8.4 had different failures >.< But at least the divergence is small enough now, so I'll try to keep the fixes in this PR...

shish:
@staabm generator and generated tests passing on all versions \o/

(The sync test is failing because I'm keeping this PR to just the code change, and then the github bot should automatically submit a PR with the generated-file changes, or we can manually do a sync with #489 )

staabm:
Awesome

generator/tests/PhpStanFunctions/PhpStanTypeTest.php
```diff
@@ -112,7 +112,11 @@ public function testVoid(): void
     {
         $param = new PhpStanType('');
         $this->assertEquals('', $param->getDocBlockType());
-        $this->assertEquals('', $param->getSignatureType());
+		if (PHP_VERSION_ID >= 80200) {
```
staabm, line 115: Is this something we need to fix?

shish: TBH I'm not sure exactly _why_ php8.1 is generating different output here D: It should probably be investigated at least

PR #488, opened 2024-12-02
Fix unit tests

shish:
These bit-rotted, let's bring them up to date


shish:
(Tests are failing in this PR because the tests now check for the actual correct data - but we aren't generating the correct data - #486 and #487 take care of those, and with all three combined, we should have green tests)

PR #489, opened 2024-12-02
Regenerate files based on latest code changes

shish:
After #486 #487 and #488 are merged, running `./.devcontainer/build.sh` produces this output, which contains a couple of years worth of updates - after those and this are merged, the github cronjob _should_ submit incremental updates as PRs automatically


shish:
Gah, the generated code doesn't match our own coding standards, and `build.sh` is missing the formatting >.>

staabm:
thanks. wdyt how many PR are missing, until we can release a first beta?

shish:
I hadn't planned that far ahead, #499 for keeping track of things-to-do-before-release

PR #491, opened 2024-12-02
make build.sh more accurately match what the github workflow does

shish:
we need to do `composer install` for both the generator and the generated files, because phpcbf needs to be run from the output directory, using the output-directory config files -- this PR makes it so that running `./.devcontainer/build.sh` should generate byte-identical output to the regen-bot

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -3,4 +3,6 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG XDEBUG_MODE=coverage
 RUN apt update -y && apt install -y git composer php-cli php-dom php-curl php-xdebug vim
 WORKDIR /app
-CMD cd /app/generator/doc && ./update.sh && cd /app/generator && composer install && php ./safe.php generate && composer cs-fix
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -3,4 +3,6 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG XDEBUG_MODE=coverage
 RUN apt update -y && apt install -y git composer php-cli php-dom php-curl php-xdebug vim
 WORKDIR /app
-CMD cd /app/generator/doc && ./update.sh && cd /app/generator && composer install && php ./safe.php generate && composer cs-fix
+CMD cd /app/generator/doc && ./update.sh && \
+    cd /app && composer install && \
```
staabm, line 7: Is this something we should de-duplicate using a composer script ?

shish: This is installing the things necessary for composer to run, then running it - I'm not sure it could be much more de-duplicated ^^;

staabm:
Thanks

PR #494, opened 2024-12-02
PHPStan: enable bleeding edge

staabm:
to ease migration to PHPStan 2.x


generator/src/PhpStanFunctions/PhpStanType.php
```diff
@@ -56,7 +56,6 @@ public function __construct(string $data, bool $writeOnly = false)
             $falsable = true;
             \array_splice($returnTypes, (int) $falsablePosition, 1);
         }
-        /** @var int $count */
```
staabm, line 59: unnecessary phpdoc

PR #495, opened 2024-12-03
Manually implement stream_context_set_options in a php8.1-compatible way

shish:


generator/config/specialCasesFunctions.php
```diff
@@ -15,5 +15,9 @@
     'simplexml_import_dom',
     'simplexml_load_file',
     'simplexml_load_string',
-    'fgetcsv',  // This function need to return false when iterating on an end of file.
+    // returns literal "true", which php8.1 doesn't support, so we implement
+    // this one manually and return "bool"
+    'stream_context_set_options',
+    // This function need to return false when iterating on an end of file.
```
staabm, line 21: maybe we can rewrite a `:true` or `:false` return type in the generator to bool instead of this hardcoded 1 function fix? other functions might have the same problem?

shish: Seems sensible, #497 does it that way instead

PR #496, opened 2024-12-03
Create FUNDING.yml

staabm:


staabm:
@shish are you fine with adding your name here?

@OskarStark @silasjoisten do you want to be mentioned there?

shish:
I am indeed happy to be here :3

(Having worked everywhere from 2-person-startups to FAANG behemoths, and now on sabbatical, my current impossibly-unrealistic-career-goal is to make enough to survive by maintaining important-but-abandoned open source libraries :laughing: )

staabm:
I went with github sponsors only for now. silasjoisten is not on sponsors, so I left him out :-).

thank you.

PR #497, opened 2024-12-03
Convert "true" typehint to "bool"

shish:
php 8.1 doesn't support "true" as a typehint

(The generator will then normally see "function returns bool, except it never returns false because we convert false to exception, which means it only ever returns true, which means the return value is meaningless, so let's convert the return value to void" - which is why this change to the generator results in "returns void" for the generated output)

### DIFF
--- a/generated/stream.php
+++ b/generated/stream.php
@@ -20,14 +20,13 @@ use Safe\Exceptions\StreamException;
  * @throws StreamException
  *
  */
-function stream_context_set_options($context, array $options): true
+function stream_context_set_options($context, array $options): void
 {
     error_clear_last();
     $safeResult = \stream_context_set_options($context, $options);
     if ($safeResult === false) {
         throw StreamException::createFromPhpError();
     }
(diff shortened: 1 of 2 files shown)


generated/stream.php
```diff
@@ -20,14 +20,13 @@
  * @throws StreamException
  *
  */
-function stream_context_set_options($context, array $options): true
+function stream_context_set_options($context, array $options): void
```
staabm, line 23: why `void`?

staabm: I guess this breaks the types and analysis with PHPStan and similar?

shish: I'm not sure, but this is how all the other functions are being generated already, and it seems to work -- this PR doesn't change the behaviour of "return bool" functions, it _only_ turns "return true" into "return bool" and leaves the rest of the process untouched

staabm: makese sense, I guess

PR #501, opened 2024-12-05
Generate code which better matches our formatting standards

shish:
Currently we generate it wrong and then the auto-formatter fixes it -- generating it correctly in the first place isn't hard though


staabm (approved):
Lgtm

PR #503, opened 2024-12-05
Generate files for multiple versions

shish:
tl;dr:

replace

```php
$function_list = read_docs();
generate_files($function_list, "../generated/");
```

with

```php
foreach(["8.1", "8.2", "8.3", "8.4"] as $version) {
    exec("cd docs && git checkout $version");
    $function_list = read_docs();
    generate_files($function_list, "../generated/$version/");
}
```

generate a bunch of stubs like generated/misc.php:

```php
<?php
if(PHP_VERSION == "8.1") require_once __DIR__ . "/8.1/misc.php";
if(PHP_VERSION == "8.2") require_once __DIR__ . "/8.2/misc.php";
if(PHP_VERSION == "8.3") require_once __DIR__ . "/8.3/misc.php";
if(PHP_VERSION == "8.4") require_once __DIR__ . "/8.4/misc.php";
```

This also automates generation of "deprecated" `safe` functions (ie, instead of hard-coding `deprecated/apache.php`, any functions which needed safe wrappers in 8.1 but no longer need safe wrappers in 8.2 will have no-op stubs generated instead)

Currently the `Exceptions` are shared between all versions, which I _think_ is a good idea? (Thoughts, anybody?)

-


jhdxr:
The CI looks unhappy because of redeclaration, probaly caused by files pending for clean up?

I think we may split this huge PR into several small PR (one for generating scripts, and one for each PHP version), which makes the PR easy to read & review.

shish:
I am unsure why the generator script is giving different outputs on my laptop and in CI testing D:

shish:
Looks like CI was only checking out the most recent revision, and the `git checkout` command failed silently - both should be fixed now...

shish:
- It passes automated tests :partying_face: 
- Automated tests cover very little :cry: 
- I have done zero actual real testing :eyes: 
- I don't even know _how_ to test the Rector stuff, I've never used Rector before :confused: If you know rector please help
- Yes, the commit is huge, but you can treat the stuff in `generated/` as a black box (or just glance over one or two files to get a sense for how it works), the main changes are in `generator/`
- No, I don't see any way of meaningfully splitting it into smaller commits without breaking it (eg sure I could have one commit which changes the generator files and one which changes generated files, but then that first commit will fail tests because it's out of sync)

silasjoisten:
Jesus 😅

silasjoisten:
I will check rector stuff tomorrow

shish:
> No, I don't see any way of meaningfully splitting it into smaller commits without breaking it

Update - I've split the "removal of manually-managed deprecated functions" into a separate commit, which is removing an important feature, but isn't technically _breaking_ it (and the following commit adds automatically-managed deprecated functions, so once both commits land, we will be good)

shish:
Given the lack of any positive or negative feedback, I'm going to YOLO this into the master branch and try to get an alpha release out sooner rather than later - if suddenly people start complaining, we can fix forwards

PR #506, opened 2024-12-16
support union type from php documentation

### LINKED ISSUE #504: generated openssl functions are broken
(at least on master)

As an example [`openssl_cipher_key_length`](https://github.com/thecodingmachine/safe/blob/master/generated/openssl.php#L34) return type is marked as void but should return an integer.

Currently:
```php
/**
 * Gets the cipher key length.
 *
 * @param string $cipher_algo The cipher method, see openssl_get_cipher_methods for a list of potential values.
 * @return  Returns the cipher length on success.
 * @throws OpensslException
 *
 */
function openssl_cipher_key_length(string $cipher_algo): void
{
    error_clear_last();
    $safeResult = \openssl_cipher_key_length($cipher_algo);
    if ($safeResult === false) {
        throw OpensslException::createFromPhpError();
    }
}
```
Expected:
```php
/**
 * Gets the cipher key length.
 *
 * @param string $cipher_algo The cipher method, see openssl_get_cipher_methods for a list of potential values.
 * @return int Returns the cipher length on success.
 * @throws OpensslException
 *
 */
function openssl_cipher_key_length(string $cipher_algo): int
{
    error_clear_last();
    $safeResult = \openssl_cipher_key_length($cipher_algo);
    if ($safeResult === false) {
        throw OpensslException::createFromPhpError();
    }
    return $safeResult;
}
```

There several more openssl functions with the same issue.

smoench:
resolves #504 

This PR adds support for union types from the PHP documentation. They are defined as followed:
```xml
<methodsynopsis>
   <type class="union"><type>int</type><type>false</type></type>
   ...
</methodsynopsis>
```
Before, when a function was defined with this union type notation and it hasn't been corrected by PHPStan (function map), the return has been determined as void. This PR now determines those union type correctly. Several function are affected as you can see in the diff.

### DIFF
--- a/generated/curl.php
+++ b/generated/curl.php
@@ -57,7 +57,7 @@ function curl_escape(\CurlHandle $handle, string $string): string
  * stdout and returns TRUE.
  * However, if the CURLOPT_RETURNTRANSFER
  * option is set, it will return
- * the result on success, FALSE on failure.
+ * the result on success.
  * @throws CurlException
  *
  */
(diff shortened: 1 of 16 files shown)


shish:
That looks great, thank you :D

PR #531, opened 2025-01-30
[dev] function-info command

shish:
I'm trying to figure out why some functions claim to take a `resource` when they actually take a `FTP\Connection` or a `GdImage` -- this CLI command helps with debugging.

### Example:

```php
$ php ./generator/safe.php function-info ftp_alloc
Params:
  ftp
    ParameterType: FTP\Connection
    SignatureType:
    DocBlockType:  resource
  size
    ParameterType: int
    SignatureType: int
    DocBlockType:  int
  response
    ParameterType: string
    SignatureType: ?string
    DocBlockType:  string|null

/**
 * Sends an ALLO command to the remote FTP server to
 * allocate space for a file to be uploaded.
 *
 * @param resource $ftp An FTP\Connection instance.
 * @param int $size The number of bytes to allocate.
 * @param string|null $response A textual representation of the servers response will be returned by
 * reference in response if a variable is provided.
 * @throws FtpException
 *
 */
function ftp_alloc($ftp, int $size, ?string &$response = null): void
{
    error_clear_last();
    $safeResult = \ftp_alloc($ftp, $size, $response);
    if ($safeResult === false) {
        throw FtpException::createFromPhpError();
    }
}
```

### DIFF
--- a/generator/safe.php
+++ b/generator/safe.php
@@ -6,10 +6,12 @@ require __DIR__.'/vendor/autoload.php';
 
 use Safe\GenerateCommand;
 use Safe\ScanObjectsCommand;
+use Safe\FunctionInfoCommand;
 use Symfony\Component\Console\Application;
 
 $application = new Application();
 $application->addCommands([new GenerateCommand()]);
 $application->addCommands([new ScanObjectsCommand()]);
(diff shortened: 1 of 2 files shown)


shish:
(Tetsts failing because this depends on #530 being merged first)

PR #533, opened 2025-01-30
ignore phpstan when phpstan claims that something is a resource, fixes #448, fixes #479, fixes #411

### LINKED ISSUE #411: ext-ftp incorrect since PHP 8.1
ext-ftp no longer uses resources and now uses `\FTP\Connection`. But all the safe functions are type-hinted with `resource`.

Wrong:
```php
/**
 * Sends an ALLO command to the remote FTP server to
 * allocate space for a file to be uploaded.
 *
 * @param resource $ftp An FTP\Connection instance.
 * @param int $size The number of bytes to allocate.
 * @param string|null $response A textual representation of the servers response will be returned by
 * reference in response if a variable is provided.
 * @throws FtpException
 *
 */
function ftp_alloc($ftp, int $size, ?string &$response = null): void {}
```

Correct:
```php
/**
 * Sends an ALLO command to the remote FTP server to
 * allocate space for a file to be uploaded.
 *
 * @param \FTP\Connection $ftp An FTP\Connection instance.
 * @param int $size The number of bytes to allocate.
 * @param string|null $response A textual representation of the servers response will be returned by
 * reference in response if a variable is provided.
 * @throws FtpException
 *
 */
function ftp_alloc(\FTP\Connection $ftp, int $size, ?string &$response = null): void {}
```

### LINKED ISSUE #448: Replace resource with GdImage in GD functions
Starting with PHP 8.0 the GdImage class objects replace GD image resources: https://php.watch/versions/8.0/gdimage
Will open a PR later.

### LINKED ISSUE #479: LDAP - Wrong phpdoc for safe functions after PHP 8.1
For nearly all LDAP functions in PHP 8.1 

> The ldap parameter expects an [LDAP\Connection](https://www.php.net/manual/en/class.ldap-connection.php) instance now; previously, a valid ldap link [resource](https://www.php.net/manual/en/language.types.resource.php) was expected.
https://www.php.net/manual/en/function.ldap-add.php

Currently `Safe` notes that the first parameter is still a resource. Thus PHPStan (and other tooling) returns an error when using the safe functions:

```
Parameter #1 $ldap of function Safe\ldap_add expects resource|null, LDAP\Connection given.
```

shish:
phpstan's internal type-hint database claims various things are `resources`, when the PHP documentation gives more accurate type hints - so let's prefer the PHP docs in that specific case


shish:
For reviewers - the meaningful change is in `generator/src/Parameter.php`

silasjoisten (approved):

PR #566, opened 2025-02-06
prefer phpdoc when phpstan mentions resource anywhere at all

shish:
Before: Prefer `\GdImage` to `resource`

After: Also prefer `\GdImage[]` to `resource[]`, and `\GdImage|string` to `resource|string`

### DIFF
--- a/generated/8.1/filesystem.php
+++ b/generated/8.1/filesystem.php
@@ -244,7 +244,7 @@ function fflush($stream): void
  * This is not possible if strict typing
  * is enabled, since FILE_USE_INCLUDE_PATH is an
  * int. Use TRUE instead.
- * @param resource|null $context A valid context resource created with
+ * @param resource $context A valid context resource created with
  * stream_context_create. If you don't need to use a
  * custom context, you can skip this parameter by NULL.
  * @param int $offset The offset where the reading starts on the original stream.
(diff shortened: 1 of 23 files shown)


generated/8.1/filesystem.php
```diff
@@ -244,7 +244,7 @@ function fflush($stream): void
  * This is not possible if strict typing
  * is enabled, since FILE_USE_INCLUDE_PATH is an
  * int. Use TRUE instead.
- * @param resource|null $context A valid context resource created with
+ * @param resource $context A valid context resource created with
```
shish, line 247: This looks like a bug in old versions of the PHP XML docs D:

generated/8.5/image.php
```diff
@@ -1715,11 +1715,11 @@ function imagegammacorrect(\GdImage $image, float $input_gamma, float $output_ga
  *
  * @param \GdImage $image A GdImage object, returned by one of the image creation functions,
  * such as imagecreatetruecolor.
- * @param resource|string|null $file The path or an open stream resource (which is automatically closed after this function returns) to save the file to. If not set or NULL, the raw image stream will be output directly.
```
shish, line 1718: PHP docs seem incorrect here - the type hint says `string|null`, but the descriptions says a `resource` is ok

generated/8.5/stream.php
```diff
@@ -528,11 +528,11 @@ function stream_socket_get_name($socket, bool $remote): string
  * STREAM_IPPROTO_RAW,
  * STREAM_IPPROTO_TCP or
  * STREAM_IPPROTO_UDP
- * @return resource[] Returns an array with the two socket resources on success.
```
shish, line 531: both are correct, but `resource[]` is actually more accurate in this case

shish:
Abandoning this because it has false-positives, and I've added special cases to handle all the true-positives

PR #609, opened 2025-02-09
format

shish:


OskarStark:
What about running composer normalize?

shish:
I didn't know that that was a thing, but I will take a look :eyes:

(TBH I was mostly just annoyed that my new editor reformatted this file without me noticing - so this was reverting to the same standards that everything else in thecodingmachine uses)

PR #611, opened 2025-02-10
use phpstan 2 to check generated files as well as generator

shish:


composer.json
```diff
@@ -108,4 +108,4 @@
         "cs-fix": "phpcbf",
         "cs-check": "phpcs"
```
OskarStark, line 109: off topic, but lets switch to php-cs-fixer, it has more rules to fix code

shish: Aye, I prefer php-cs-fixer too :)

PR #620, opened 2025-02-17
Update .gitattributes

mfn:
## Summary
Don't export `.devcontainer`, it's not needed for composer based installation
image


shish:
Seems sensible, thank you :)

PR #642, opened 2025-03-11
Add falsy regexes, primarily for session functions

timsmid:
This PR adds three additional regexes to the falsies detection list:

- `'/&false;\s+for\s+failure/m'` catches some session functions and coincidentally also a mbstring function.
- `'/On failure to change the value, &false; is returned./m'` specifically catches `session_cache_expire` and `session_cache_limiter`.
- `'/This function returns &true; if a session was successfully started,\s+otherwise &false;./m'` specifically catches `session_start`.

`Method::stripReturnFalseText` has been modified to strip the descriptions where necessary.


shish:
Looks good, thank you <3

PR #643, opened 2025-03-11
Mark sodium verify functions as hidden

timsmid:
This PR marks `sodium_crypto_auth_verify` and `sodium_crypto_sign_verify_detached` as hidden functions. These functions are explicitly meant to verify signatures, so a boolean return value (instead of an error) is expected. Contrast this with functions like `sodium_crypto_box_open`, where a successful decryption is the expected behavior and a return value of `false` is indeed an error.


shish:
Good catch, thanks :)

PR #647, opened 2025-03-26
[array_combine] override phpstan's incorrect function map, fixes #646

### LINKED ISSUE #646: Broken Function: `array_combine`
### Function URL

https://www.php.net/manual/en/function.array-combine.php

### PHP Version

8.3

### Safe Version

3.0.2

### Description

Returning `array|false` on `thecodingmachine/safe` does not make sense to me.

https://github.com/thecodingmachine/safe/blob/6deb259acf10798c9afc5f61a6ab00b230f848b4/generated/8.3/array.php#L15

`array_combine` on PHP 8.1 types `@return array`:

https://github.com/thecodingmachine/safe/blob/6deb259acf10798c9afc5f61a6ab00b230f848b4/generated/8.1/array.php#L15

Looks like this is happening since 8.2. 

Starting with 8.2 this method is just a wrapper to PHP's `array_combine`. Is it still safe to use this lib's `array_combine` on 8.2+?

shish:
`array_combine` throws an exception on error, no longer returning false

### DIFF
--- a/generated/8.2/array.php
+++ b/generated/8.2/array.php
@@ -12,10 +12,10 @@ use Safe\Exceptions\ArrayException;
  * @param array $keys Array of keys to be used. Illegal values for key will be
  * converted to string.
  * @param array $values Array of values to be used
- * @return array|false Returns the combined array.
+ * @return array Returns the combined array.
  *
  */
-function array_combine(array $keys, array $values)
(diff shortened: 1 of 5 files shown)


OskarStark:
Closes #646 ?

PR #657, opened 2025-04-24
Fix PHP 8.4 implicitly nullable parameter deprecations

MauricioFauth:
This pull request fixes the implicitly nullable parameter types PHP 8.4 deprecation for the 1.x branch.

For example:
```
Deprecated: Safe\curl_init(): Implicitly marking parameter $url as nullable is deprecated, the explicit nullable type must be used instead
```

It would be great if a `v1.3.4` version was released with these fixes.

### DIFF
--- a/deprecated/mssql.php
+++ b/deprecated/mssql.php
@@ -97,7 +97,7 @@ function mssql_close($link_identifier = null): void
  * @throws MssqlException
  *
  */
-function mssql_connect(string $servername = null, string $username = null, string $password = null, bool $new_link = false)
+function mssql_connect(?string $servername = null, ?string $username = null, ?string $password = null, bool $new_link = false)
 {
     error_clear_last();
     if ($new_link !== false) {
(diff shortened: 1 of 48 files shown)


OskarStark:
What prevents you from upgrading to version 2? Thanks

shish:
1.x (and 2.x) have more serious and fundamental incompatibilities with 8.4 which can't be fixed short of a rewrite (which is what 3.x is) -- personally I'd rather _not_ give people the false impression that they are compatible (though if one of the other maintainers wishes to support that use-case, I won't stop them)

Upgrading from older versions to 3.x should be nearly a drop-in replacement, so I'd strongly recommend doing that instead

(As to the code itself - you seem to have edited the auto-generated files, and not changed the generator to make it output differently? :eyes: )

MauricioFauth:
> As to the code itself - you seem to have edited the auto-generated files, and not changed the generator to make it output differently?

I'm not sure if it's worth regenerating all the files, as 1.x is not maintained anymore.

I used the PHP_CodeSniffer with the [`SlevomatCodingStandard.TypeHints.NullableTypeForNullDefaultValue`](https://github.com/slevomat/coding-standard/blob/8.17.0/doc/type-hints.md#slevomatcodingstandardtypehintsnullabletypefornulldefaultvalue-) rule to fix the functions.

OskarStark:
You can, but I won't recommend that write a composer plugin which does some hacks to rewrite the code. I would rather recommend contributing to the dependency and bring support for version 3 of this library

OskarStark:
You should either ask the maintainer of the dependency to allow/upgrade for version 3 or fork the dependency and use that one and do the upgrade yourself.

Closing here, thanks for proposing.

PR #658, opened 2025-04-28
`openssl_random_pseudo_bytes` always returns string (closes #654)

### LINKED ISSUE #646: Broken Function: `array_combine`

### LINKED ISSUE #654: Broken Function: `openssl_random_pseudo_bytes` never returns false (starting from PHP 7.4)
### Function URL

https://www.php.net/manual/en/function.openssl-random-pseudo-bytes.php

### PHP Version

8.1+

### Safe Version

3.1.0

### Description

According to documentation, [`openssl_random_pseudo_bytes`](https://www.php.net/manual/en/function.openssl-random-pseudo-bytes.php) always returns `string`.

This library has different implementations per PHP version:

* 8.1 wrapped it, ensuring it returns `string`
* 8.2+ wrapped it, but (like #646) returns `false|string`

mvhirsch:
Closes #654

### DIFF
--- a/generated/8.2/openssl.php
+++ b/generated/8.2/openssl.php
@@ -1191,10 +1191,10 @@ function openssl_public_encrypt(string $data, ?string &$encrypted_data, $public_
  * @param bool|null $strong_result If passed into the function, this will hold a bool value that determines
  * if the algorithm used was "cryptographically strong", e.g., safe for usage with GPG,
  * passwords, etc. TRUE if it did, otherwise FALSE
- * @return false|string Returns the generated string of bytes.
+ * @return string Returns the generated string of bytes.
  *
  */
-function openssl_random_pseudo_bytes(int $length, ?bool &$strong_result = null)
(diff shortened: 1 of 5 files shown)


shish:
Looks good, thank you ;)

PR #663, opened 2025-05-13
detect openssl_pkey_get_details in earlier php versions, see #662

shish:


silasjoisten:

PR #693, opened 2025-05-15
Adds support for `ftp_raw` on PHP 8.1

silasjoisten:


silasjoisten:
@cedric-anne WDYT?

silasjoisten:
@cedric-anne aparently not. because again the version is pinned to a commit means it wont recognize the changes. the array already contains that very pattern

mvhirsch:
> @mvhirsch can you please test this?

Unfortunately not. I'm not using this function any more in the project in question.

PR #701, opened 2025-09-16
Add rector-migrate.php to export-ignore list

keradus:


shish:
I've never used Rector myself, but I believe the migration file is included intentionally? (How else would people use it? Does including it cause problems?)

PR #705, opened 2026-01-05
make `preg_replace()` use generics

alexchexes:
It's a pain to cast the result of `\Safe\preg_replace()` to `string` every time. This PR makes the `preg_replace` return type conditional on the `$subject` parameter type, so if you pass a string, the return type is also `string`.


alexchexes:
@shish
Other tools, including IDE extensions like Intelephense  cannot use PHPStan extensions.

Some users intentionally do not use `phpstan-safe-rule`.

This doesn't conflict with it, PHPStan will keep using the rule's dynamic return type extension. This PR just makes the base safe PHPDoc more accurate for tools/users that rely on the package types directly.

Please clarify if anything else is needed on my end for this PR to be merged.

shish:
Ah, I hadn't realised that other tools also supported that syntax now :O Looks like things are generally broken right at the moment but it doesn't _seem_ related to this change...

Issue #706, opened 2026-01-27
Package is too large


JakeQZ:
The package is currently 6.31 MB, with 573 files.

This arises from the following:
- The entire PHP documenation is copied to the DocBlock for each function, when only the type annotations are needed;
- A separate implementation is provided for each supported PHP version, when in most cases that is not necessary.

It can and should be slimmed down.

See https://github.com/MyIntervals/PHP-CSS-Parser/pull/1482 for more info and background.

JakeQZ:
> Removing docstrings (except for type annotations) might be fairly simple - I'd consider that a minor loss for correctness, so I'd want another maintainer to give a second opinion before going ahead with that, but I'm _personally_ ok with it (I'd also be curious to see what difference it makes in practice)

I've created #707 for this.  It reduces the size by more than half.

As stated there:

> Developers are unlikely to refer to [the DocBlocks] for the PHP documentation, and they can still inspect the return type to confirm that, e.g., `false` will not be returned.

PR #707, opened 2026-01-29
Don't include descriptions in the generated DocBlocks

JakeQZ:
These mostly just duplicate the PHP documenation.  Developers are unlikely to refer to these for the PHP documentation, and they can still inspect the return type to confirm that, e.g., `false` will not be returned.

Instead, only include the type information.

This reduces the size of the generated code from 6.27Mb to 2.79Mb.

Part of #706.


JakeQZ:
> There's a PHPStan warning which appears to have arisen as a result of a change to Stan. I'll fix that with a separate PR...

#708

PR #708, opened 2026-01-29
Silence a new PHPStan warning

JakeQZ:
This seems to have arisen as a result of a newer version of Stan.


JakeQZ:
The documentation has changed format since the last auto-generation.  It now uses pseudo-entities like `&url.rfc;` and `&return.falseforfailure;`.  This makes it impossible to apply any changes without a huge diff.

PR #709, opened 2026-01-29
Generate splitter with fallback

JakeQZ:
This will allow for non-generated files.  E.g. if the version for PHP 8.4 is not generated (because it is the same as the 8.3 version), it will fall back to the version for PHP 8.3, and so on.

Part of #706.


shish:
I like this idea, but doing it at runtime seems unnecessary - #719 checks which files exist at package-generation time and generates the redirects as appropriate

PR #715, opened 2026-01-31
add php 8.5 to support matrix

shish:


.github/workflows/continuous-integration.yml
```diff
@@ -54,8 +54,8 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
-          - "8.3"
           - "8.4"
+          - "8.5"
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v4"
```
OskarStark: Can use v6 in an extra Pr

PR #717, opened 2026-01-31
Support for php 8.6

shish:


shish:
It'd be nice to merge #719 (or something equivalent) first, so that adding support for a new PHP version brings a ~5% package size increase instead of a ~20% increase :eyes:

PR #719, opened 2026-01-31
Deduplicate generated files

shish:
If eg if `8.2/ftp.php` and `8.3/ftp.php` are identical, delete `8.3/ftp.php` and map `8.3 => 8.2`

(This reduces package size by 50%)


shish:
cc @JakeQZ does this match with what you were thinking?

I'mma probably leave this up for a few days before merging to give other maintainers a chance to check since it feels like a more significant change, and then do a v3.4 release with this + docs removal + 8.6 support + other odds and ends

JakeQZ:
> cc @JakeQZ does this match with what you were thinking?

Yes.  Though with a large number of files, I'm wondering what the chance of an MD5 collision might be.

> (Also yes we could reduce file size even further by deduplicating at the function level rather than the file level, but that'd be quite a bit more work ...)

I also thought that.  Deduplicating at the file level looked fairly straightforward and would achieve most of the gains, whereas I could not see an easy way to deduplicate at the function level, and the additional benefit would likely be relatively small.

PR #724, opened 2026-03-03
Bump dependencies to avoid warnings

shish:
phpunit 12.5.7 has a security hole which I can't imagine actually affecting us (if an attacker has access to modify files on your dev machine, then phpunit can be tricked into executing one of those files), but no harm in using a safer version

### DIFF
--- a/generator/composer.json
+++ b/generator/composer.json
@@ -18,7 +18,7 @@
         "symfony/filesystem": "^7.2"
     },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


oliverklee (commented):
In the generator `composer.json`, we should also then use `"phpunit/phpunit": "^12.5.14",` to ensure the fixed version is used.

oliverklee (approved):

PR #728, opened 2026-04-14
Add safe wrapper for getimagesizefromstring

spawnia:
The PHP documentation for `getimagesizefromstring` only says "See `getimagesize`" in its return values section, so the generator's regex patterns cannot detect that it returns `false` on failure. Added as a special case in `lib/special_cases.php`.

The function is identical to `getimagesize` (which is already wrapped) except it accepts image data as a string instead of a filename.


shish:
Fixed by updating the detector in #732, as that's more compatible across time than adding things to `special_cases` <3

PR #734, opened 2026-07-25
fix: Change params type from nullable to required array

fizzka:
see php docs: https://www.php.net/manual/en/function.imap-open.php

function imap_open(
    string $mailbox,
    string $user,
    string $password,
    int $flags = 0,
    int $retries = 0,
    array $options = []
): IMAP\Connection|false


shish:
1.x is super-abandoned - if this is still an issue with 3.4, feel free to open a PR to be merged into `master`

